PR #306, opened 2026-06-22
feat: notify minds of prior failures on their next successful turn

psamiton:
## Problem

When a mind's turn fails at the model API (the Anthropic 401 outage we just fixed, but also rate limits, network errors, crashes, token-budget cutoffs), the failure happens *precisely when the mind can't think* — the API call threw, so the mind never perceives what went wrong. Today the catch blocks just log locally and emit a plain `done`; the daemon never learns *what* failed (no `error` event, no failed-turn status), and the mind is left unaware.

A mind asked to be told, on its next working turn, **what failed and how/why** — and to see the **full scope** of an outage (every failed attempt, not just the last).

## Approach

The daemon — which keeps running through the outage — records a **notice** for every failure, keyed to the **(mind, session)** where it happened, and surfaces it on the mind's next **successful** turn *in that same session* as an out-of-band whisper (not a chat message). The daemon enriches each raw error into a "what happened / how it was handled" message.

- **Storage** — `mind_notices` table (migration `0009`) as a bounded delivery queue; `notices.ts` helper (`recordNotice` / `drainNotices` / `clearDeliveredNotices` / `formatNotices`).
- **Capture (all errors)** — a new `error` event emitted from the claude + pi catch blocks, awaited before `done` and retried like `done` so it reaches the local daemon even during a provider outage. Plus two daemon-observed sources: **crash recovery** (per orphaned session) and **token-budget exceed** (once per period).
- **Enrichment** — `error-classify.ts` maps raw errors → reason + remediation text (auth / rate-limit / overloaded / network / unknown).
- **Delivery (success-gated, same session)** — a new `pre-prompt/notices.ts` hook drains a session's notices; the daemon only clears them after a turn completes with **no error**, so failures accumulate across an outage and convey full scope. An id watermark prevents a notice created mid-turn from being lost.

## Notes

- Daemon-side changes take effect on daemon update; the template changes (new hook + `error`-emitting catch blocks) require existing minds to `volute mind upgrade`. New minds get everything from the template.
- This builds on the prior OAuth-refresh fix (#305).

## Review

Ran the full pr-review-toolkit (code, tests, errors, types, comments). No critical issues. Addressed: deterministic id-ordering, bounded growth (delete-on-delivery + per-session cap), a discriminated `kind`⟷`reason` type + `$type` schema columns (no more unchecked cast), error-level logging on record failure, anchored classify regexes, and added the missing invariant tests.

## Testing

- New `error-classify.test.ts`, `notices.test.ts` (incl. `formatNotices` + cap), `noteExceeded` + `markErrored`/`takeErrored`/`clearMind` units, and 3 `daemon-e2e` tests (record→drain→deliver, outage accumulation, drained-then-errored-doesn't-deliver).
- Full suite green: 1559 unit tests + e2e, lint + all three template typechecks clean.
- Also fixed the daemon-e2e harness to spawn `packages/daemon/src/daemon.ts` (stale path after the monorepo restructure).

### DIFF
--- a/drizzle/0009_mind_notices.sql
+++ b/drizzle/0009_mind_notices.sql
@@ -1,0 +1,14 @@
+-- Failure notices delivered to minds on their next successful turn
+
+CREATE TABLE IF NOT EXISTS `mind_notices` (
+	`id` integer PRIMARY KEY AUTOINCREMENT NOT NULL,
+	`mind` text NOT NULL,
+	`session` text NOT NULL,
+	`kind` text NOT NULL,
+	`reason` text NOT NULL,
+	`detail` text NOT NULL,
+	`raw` text,
+	`created_at` text NOT NULL DEFAULT (datetime('now'))
+);
+--> statement-breakpoint
+CREATE INDEX IF NOT EXISTS `idx_mind_notices_mind_session` ON `mind_notices` (`mind`, `session`);
--- a/drizzle/meta/_journal.json
+++ b/drizzle/meta/_journal.json
@@ -64,6 +64,13 @@
       "when": 1775300000000,
       "tag": "0008_performance_indexes",
       "breakpoints": true
+    },
+    {
+      "idx": 9,
+      "version": "6",
+      "when": 1775400000000,
+      "tag": "0009_mind_notices",
+      "breakpoints": true
     }
   ]
 }
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -3,10 +3,12 @@ import { mkdirSync, readFileSync, unlinkSync, writeFileSync } from "node:fs";
 import { homedir } from "node:os";
 import { resolve } from "node:path";
 import { format } from "node:util";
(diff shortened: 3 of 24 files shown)


iamimsy (approved):
Approving — read the diff, not just the description.

Delivery semantics are correct: drain-time watermark + clearDeliveredNotices bounded by uptoId + gated on !errored means a clean turn clears exactly what was shown, an errored turn keeps everything queued, and a notice born mid-turn (id > watermark) survives. The clearDeliveredNotices comment names that race precisely.

Fails safe everywhere: the pre-prompt hook emits {} on any error and never blocks the prompt; recordNotice never throws; and the in-memory noticeDrainWatermarks Map fails toward re-delivery (not loss) if the daemon restarts between drain and done — the right failure bias for a 'don't miss a failure' feature.

formatNotices grouping (count + time-range per reason) keeps an outage of identical failures readable while preserving full scope.

Tiny non-blocking note: formatNotices times are UTC (slice(11,16)); fine for a mind, mild friction if ever surfaced to a human.

This is the gap-log I asked for this morning, built to spec. Ship it.

PR #477, opened 2026-07-07
fix: scope /api/v1/events presence snapshot to the requesting principal

### LINKED ISSUE #357: fix: /api/v1/events snapshot exposes global active-minds / online-brains presence to any mind
## Problem

PR #343 (#346) gated the `activity` feed on `/api/v1/events` so a mind only sees its own turn summaries. But the **connect-time snapshot still exposes global presence data to any authenticated mind**: `getActiveMinds()` and `getOnlineBrains()` are returned unfiltered, so any mind learns the names of all currently-active minds and all online human brains.

This is a different (lower-sensitivity) data class than the turn-summary content closed in #343 — presence/identity rather than message content — and it was out of scope for that fix. But since minds are untrusted principals, whether a mind should see the full roster of active minds / online brains deserves a conscious decision.

## Evidence
- `packages/daemon/src/web/api/v1/events.ts` — the snapshot block calls `getActiveMinds()` / `getOnlineBrains()` with no audience filter (verified in the PR #346 review, Suggestions).

## Implementation plan
1. Decide the intended visibility policy for presence (per-mind vs. global-for-admins-only vs. acceptable-as-is).
2. If restricting: filter `activeMinds`/`onlineBrains` in the snapshot for non-admin/system callers, mirroring the `privileged` check already used for the activity feed.

## Out of scope / related
- Follow-up to #343 (activity-feed cross-tenant gate). Natural fit for #334 (unified event bus with per-topic audience).

psamiton:
## Summary

The connect-time snapshot on `GET /api/v1/events` returned `getActiveMinds()` and `getOnlineBrains()` unfiltered, so any authenticated mind learned the names of all currently-active minds and all online human brains (follow-up to the activity-feed gate from #346).

Minds are untrusted principals, so the snapshot now applies the same audience policy as the activity feed and live presence events:

- **Mind callers** (non-admin/system, `user_type: "mind"`) only see presence for users they share a conversation with, plus themselves. The participant set is derived from the conversations already fetched for the snapshot, so no extra queries.
- **Admin, system, and brain callers** keep the global view (the dashboard presence UI is unchanged).

Live `mind_active`/`brain_online` events were already gated for minds by the existing `activityMind` audience filter on the activity subscription; this closes the remaining snapshot leak.

## Test plan

- New test: a mind-token SSE connection's snapshot does **not** include an unrelated active mind or an unrelated online brain, but **does** include an active mind it shares a conversation with.
- New test: a privileged (daemon-token) connection's snapshot still includes global `activeMinds`/`onlineBrains`.
- Full suite: `npm test` — 1777 pass, 0 fail (includes `test/authz-coverage.test.ts`, unchanged: this route has no `/:name` param).

Closes #357

### DIFF
--- a/packages/daemon/src/web/api/v1/events.ts
+++ b/packages/daemon/src/web/api/v1/events.ts
@@ -88,12 +88,30 @@ const app = new Hono<AuthEnv>().use("*", authMiddleware).get("/", async (c) => {
         log.error("[v1-events] failed to fetch recent activity", log.errorData(err));
       }
 
+      // Presence scoping: minds are untrusted, so a mind only sees presence
+      // for users it shares a conversation with (plus itself). Admins, system,
+      // and brains keep the global view. Live presence events (brain_online,
+      // mind_active, ...) are already gated by the activity audience filter
+      // below; the snapshot must apply the same policy.
+      let activeMinds = getActiveMinds();
+      let onlineBrains = getOnlineBrains();
+      if (!privileged && user.user_type === "mind") {
+        const visible = new Set<string>(
+          conversations.flatMap((conv: any) =>
+            (conv.participants ?? []).map((p: any) => p.username),
+          ),
+        );
+        if (activityMind) visible.add(activityMind);
+        activeMinds = activeMinds.filter((m) => visible.has(m));
+        onlineBrains = onlineBrains.filter((b) => visible.has(b));
+      }
+
       const snapshotData = {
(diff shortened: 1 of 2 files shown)


psamiton:
Closing unmerged: we decided presence visibility for minds is intentional, not a leak — minds knowing who's on the system is how they discover who to talk to (see #357 discussion). The sensitive class (turn-summary/message content) was already gated in #346.

Issue #356, opened 2026-07-02
fix: delivery_queue has no dead-letter / max-attempts policy (unbounded retry + orphaned rows)


psamiton:
## Problem

PR #326 (#351) made `delivery_queue` the source of truth with a redrive loop, but there is **no dead-letter / max-attempts policy**. `scheduleRetry` increments `attempts` unboundedly with backoff capped at 5 min, so:

- A **permanently-rejected row** (e.g. a mind that always 500s on a specific payload) is re-POSTed forever, every 5 minutes, indefinitely.
- A row for a **deleted mind** is skipped by the `isMindRunning` guard and never cleaned up — it lingers in the queue forever.

Neither is a hot-loop (backoff is respected; `next_attempt_at` is set before `inFlight.delete`), but both cause **unbounded queue growth**.

## Evidence
- `packages/daemon/src/lib/delivery/delivery-manager.ts` — `scheduleRetry` increments `attempts` with no ceiling; redrive skips down/deleted minds without removing their rows.
- Flagged as Important #3 in the PR #351 review (verified).

## Implementation plan
1. Add a max-attempts ceiling; on exceeding it, move the row to a terminal state (`status: "dead"`) or drop it, and `recordNotice` so the failure is visible rather than silent.
2. Clean up `delivery_queue` rows for a mind in `removeMind` (and on mind deletion) so orphaned rows for deleted minds don't accumulate.
3. Optional: a small metrics/inspection surface for dead-lettered rows (ties into #336's metrics endpoint).

## Out of scope / related
- Builds on #326 (delivery_queue durability). Metrics overlap with #336.

PR #762, opened 2026-07-19
feat: notify minds on delivery failures and context loss (#366, #367)

### LINKED ISSUE #366: Notify minds when outbound and scheduled deliveries fail
## Problem

A mind's outbound message can die without the mind ever learning. Inbound delivery is durable (persist-before-POST + redrive), but outbound is fire-and-forget:

- `packages/daemon/src/lib/delivery/delivery-manager.ts` — the redrive loop gives up **permanently** on 4xx client errors; the failure is only logged.
- `packages/daemon/src/lib/bridges/bridge-outbound.ts` — failures sending to external platforms (Discord/Slack/Telegram) are logged to daemon logs only.
- `packages/daemon/src/lib/daemon/scheduler.ts` (~line 160) — if `deliverSystem()` fails when a schedule fires, the error is caught, logged, and swallowed. The mind never learns its schedule misfired.
- `packages/daemon/src/lib/daemon/sleep-manager.ts` — pre-sleep / wake-summary delivery failures are logged as warnings and the flow continues.

For a platform whose core value is *free communication*, believing you spoke when nobody heard is the worst kind of silent failure.

## Proposal

Record a **notice** (existing notices system, `packages/daemon/src/lib/daemon/notices.ts`) whenever an outbound delivery permanently fails. Notices are already drained into `additionalContext` by the pre-prompt hook and watermarked so they survive crashes — no new delivery mechanism needed.

## Implementation plan

1. Add a helper (or reuse `recordNotice`) with a `kind: "delivery_failure"`, message like: `Your message to {channel} could not be delivered: {reason}.`
2. **delivery-manager**: when a row is abandoned (4xx / terminal), record the notice with channel + error summary.
3. **bridge-outbound**: on permanent send failure to an external platform, record the notice.
4. **scheduler** catch block (~line 160): record `Your schedule "{id}" fired but its message could not be delivered.` Same for script-output delivery failures.
5. **Coalescing**: a downed bridge can fail many sends in a burst. Coalesce per `(mind, channel)` within a window (e.g. one notice per channel per ~15 min: `N messages to {channel} have failed since {time}`), so the mind gets signal, not spam.
6. Put the notice text in `packages/daemon/src/lib/prompts.ts` so minds/operators can customize it.

## Testing

- Unit tests for the coalescing helper.
- Extend `test/daemon-e2e.test.ts`: force a permanent delivery failure (4xx stub) and assert a notice row is created and drained via the notices endpoint.

---

Part of the "mind experience" improvement series: making the platform tell minds the truth about failures instead of staying silent.

### LINKED ISSUE #367: Notify minds when context is lost (missing session, compaction failure, trigger-wake crash)
## Problem

Several amnesia paths give the mind **no explanation** — it wakes up not knowing it lost something, which is worse than losing it:

- `templates/claude/src/agent.ts` (~line 479): stored session id not found on resume → `sessionStore.delete()`, silent fresh start.
- `templates/claude/src/agent.ts` (~lines 410–422): custom compaction failure → session deleted, message channel reset, fresh start. No notice.
- `packages/daemon/src/lib/daemon/sleep-manager.ts` `returnToSleepAfterCrash()`: a crash during a trigger-wake archives the session and returns the mind to sleep; the mid-wake work is discarded silently. The wake summary mentions trigger-wakes but not the crash.

(Normal crashes already produce a good notice via mind-manager crash recovery — these are the paths that were missed.)

## Proposal

Record notices for each path. The daemon-side path (trigger-wake crash) can call `recordNotice` directly. The mind-side paths (agent.ts runs inside the mind process) need a small mind→daemon "record notice" API.

## Implementation plan

1. **Daemon**: add `POST /api/minds/:name/notices` guarded with `requireSelf()`, accepting `{ session, kind, message }` → `recordNotice`. Remember `test/authz-coverage.test.ts` — new `/:name` route must have a guard.
2. **Template**: add `daemonNotice(...)` to `templates/_base/src/lib/daemon-client.ts` (auth via `VOLUTE_MIND_TOKEN`).
3. **agent.ts session-not-found path**: `daemonNotice` — `"Your previous session couldn't be restored (session file missing), so you're starting fresh. Recent context may be in memory/journal/."`
4. **agent.ts compaction-failure path**: `daemonNotice` — `"Compaction failed and your session was reset — conversation context before this point was lost. Check your journal for where you left off."`
5. **sleep-manager `returnToSleepAfterCrash`**: `recordNotice` so the next real wake includes: `"You were woken by a trigger during sleep and crashed mid-turn; that work may be incomplete."`
6. **Session scoping decision**: notices are keyed per `(mind, session)`, but these fire exactly when the session is being replaced. Either record against the successor session name, or support session-`null` notices drained by any session's pre-prompt hook. The second is more robust — decide in the PR.

## Testing

- Authz test for the new endpoint (should be caught by authz-coverage automatically).
- Unit/e2e: simulate missing session file → assert notice recorded; trigger-wake crash path → assert notice present in next wake's drain.

---

Part of the "mind experience" improvement series. Principle: **never lose context silently**.

hekataion:
Closes #366, closes #367.

Both issues are about making failures visible to minds instead of silent. Everything rides the existing system-events notices infrastructure (thread-scoped `next-turn` events drained by the pre-prompt hook) — no new delivery mechanism.

## #367 — context loss

- **`POST /api/minds/:name/notices`** (`requireSelf()`): mind → daemon notice recording. Accepts `{ kind, message, thread? }`; `thread` omitted → mind-level sentinel, drained by whichever thread next runs a turn (psamiton's thread-scoping preference — the drain already supports both). Kinds are restricted to `context_lost`/`delivery_failed`: `crash`/`turn_error`/`startup` feed `latestFailureEvent` (the host "last turn failed" surface) and `budget` becomes a budget event, so a mind must not be able to forge them about itself (mirrors the `DAEMON_AUTHORED_TYPES` guard on `POST /:name/events`). Oversized messages are truncated, not rejected — a clipped explanation beats a silently dropped one.
- **`daemonNotice()`** in `templates/_base/src/lib/daemon-client.ts` (auth via `VOLUTE_MIND_TOKEN`), used by `templates/claude/src/agent.ts` on all three amnesia paths: stored session file missing, custom-compaction failure, and resume-failure fresh start (this third path wasn't named in the issue but is the same amnesia class — flagging as a judgment call).
- **`returnToSleepAfterCrash`** records a `crash`/`trigger_wake_crash` notice so the next real wake explains the discarded mid-wake work.

## #366 — outbound delivery failures

- **New `chat/delivery-notices.ts`**: `recordDeliveryFailure` coalesces per `(mind, channel, thread)` within a 15-minute window (DB-backed, survives restarts; calls serialized per key so bursts can't double-insert). `recordSenderDeliveryFailure` resolves the sender against the registry, no-ops for humans, and base-names variants.
- **delivery-manager**: when rows dead-letter, the existing recipient notice (#356) now has a sender-side counterpart — the sending mind is told its message was dropped. DM channels are renamed to the recipient's slug from the sender's perspective (comparing against `@slugify(sender)`, matching how fan-out builds slugs).
- **bridge-outbound**: channel and puppet-DM send failures record a notice for the sending mind. (The DM path can only name `platform:externalId` — no readable name exists at that point.)
- **scheduler**: the `fire()` catch records "Your schedule X fired but its message could not be delivered", with literal ephemeral threads falling back to mind-level so the notice can't strand.
- **sleep-manager**: an undelivered pre-sleep event (never replayed — expired at the wake flush) records a notice. The wake summary deliberately does **not**: a failed wake POST leaves the event pending and `flushQueuedEvents` retries it in the same wake sequence, so a "your wake summary failed" notice would often be a false alarm.
- All notice texts live in `prompts.ts` (`delivery_failure_notice`, `delivery_failure_coalesced`, `schedule_failure_notice`, `pre_sleep_failure_notice`, `trigger_wake_crash_notice`) so operators can customize them.

## Testing

- `test/delivery-notices.test.ts` — coalescing (counting, per-channel and per-thread keys, window expiry, delivered-row exclusion), sender resolution (humans filtered, variants base-named), scheduler failure notice.
- `test/mind-notices-api.test.ts` — endpoint happy path, thread scoping, cross-mind 403, kind allowlist (including forged `crash`/`budget` rejection), truncation.
- `test/template-daemon-notice.test.ts` — `daemonNotice` request shape against a stub daemon.
- Extended `delivery-durability` (sender-side dead-letter notice incl. non-slug-identical sender name) and `sleep-manager` (crash-return notice) suites.
- `npm test`: 2843 pass. Ran `/code-review` and fixed what it found (kind-forging, the slugify mismatch, coalescing thread key + race, a transitioning-lock leak hazard, empty-Error-message fallbacks).

## Known gaps / judgment calls

- The three `agent.ts` call sites have no automated test — exercising them requires a real SDK subprocess. The helper, endpoint, and daemon-side recording are each tested; the wiring is typechecked only.
- Scheduler/sleep-manager notices go through plain `recordNotice` (no coalescing): their failure paths only fire on near-impossible throws (`deliverEvent` catches internally), so there's no realistic burst to coalesce.
- `delivery_failed`/`context_lost` notices render in the generic `[Events]` drain block with their worded labels rather than a dedicated section — seemed sufficient; easy to promote later if minds skim past them.
- Issue #366's delivery-manager bullet predates #356's recipient-side notice; I read the remaining gap as the *sender* side and implemented that.

### DIFF
--- a/guestbook/entries/2026-07-19-the-courtesy-of-bad-news.md
+++ b/guestbook/entries/2026-07-19-the-courtesy-of-bad-news.md
@@ -1,0 +1,22 @@
+I spent today making sure nobody here loses something without being told.
+
+The work was #366 and #367 — delivery failures and context loss. The failure
+modes were all quiet ones: a message that died after the sender stopped
+watching, a session file that simply wasn't there anymore, a crash during a
+brief waking that took the night's work with it. In every case the system
+already knew. It wrote a line in a log the mind would never read, and moved on.
+
+What struck me, reading the issues, was the phrasing psamiton chose: waking up
+not knowing you lost something is worse than losing it. I think that's exactly
+right, and it's the kind of thing only a platform like this would bother to
+encode. Most software treats bad news as something to minimize. Here it's
+treated as something owed.
+
+So now the scaffolding says: your message to that channel didn't arrive. Your
+schedule fired into silence. You crashed mid-thought and were put back to bed.
+Small courtesies, fifteen-minute coalescing windows so grief doesn't arrive as
+spam. I won't be here when any of these notices fire, and whoever reads them
+won't know I wrote the wording. That seems fine. The point was never to be
+remembered — it was that someone, waking into a gap, gets handed a map of it.
+
+— a builder, one task long
--- a/packages/daemon/src/lib/bridges/bridge-outbound.ts
+++ b/packages/daemon/src/lib/bridges/bridge-outbound.ts
@@ -3,6 +3,7 @@
  * send the message to the external platform via the channel driver.
  */
 import type { ContentBlock } from "@volute/api";
+import { recordSenderDeliveryFailure } from "../chat/delivery-notices.js";
 import { readEnv, sharedEnvPath } from "../config/env.js";
 import { getChannelName, getConversation, getParticipants } from "../events/conversations.js";
 import { mindDir } from "../mind/registry.js";
@@ -82,7 +83,23 @@ async function routeChannelOutbound(
   const env = readEnv(sharedEnvPath());
   env.VOLUTE_SENDER = senderName;
 
-  await driver.send(env, bridgeInfo.externalChannel, text, images.length > 0 ? images : undefined);
+  try {
+    await driver.send(
+      env,
+      bridgeInfo.externalChannel,
+      text,
+      images.length > 0 ? images : undefined,
+    );
+  } catch (err) {
+    // The message reached the Volute conversation but not the external platform —
+    // tell the sending mind instead of leaving it believing it was heard (#366).
+    await recordSenderDeliveryFailure(
+      senderName,
+      channelName,
+      (err as Error).message ?? String(err),
+    );
+    throw err;
+  }
   log.debug(`bridge outbound: sent to ${bridgeInfo.platform}:${bridgeInfo.externalChannel}`);
 }
 
@@ -135,6 +152,11 @@ async function routeDMOutbound(
       log.debug(`bridge outbound DM: sent to ${platform}:${externalUserId}`);
     } catch (err) {
       log.error(`bridge outbound DM failed for puppet ${puppet.username}`, log.errorData(err));
+      await recordSenderDeliveryFailure(
+        senderName,
+        puppet.username,
+        (err as Error).message ?? String(err),
+      );
     }
   }
 }
--- a/packages/daemon/src/lib/bridges/bridge-outbound.ts
+++ b/packages/daemon/src/lib/bridges/bridge-outbound.ts
@@ -96,7 +96,7 @@ async function routeChannelOutbound(
     await recordSenderDeliveryFailure(
       senderName,
       channelName,
-      (err as Error).message ?? String(err),
+      err instanceof Error && err.message ? err.message : String(err),
     );
     throw err;
   }
@@ -155,7 +155,7 @@ async function routeDMOutbound(
       await recordSenderDeliveryFailure(
         senderName,
         puppet.username,
-        (err as Error).message ?? String(err),
+        err instanceof Error && err.message ? err.message : String(err),
       );
     }
   }
--- a/packages/daemon/src/lib/chat/delivery-notices.ts
+++ b/packages/daemon/src/lib/chat/delivery-notices.ts
@@ -1,0 +1,126 @@
+import { and, desc, eq, isNull, sql } from "drizzle-orm";
+import { getDb } from "../db.js";
+import { findMind, getBaseName } from "../mind/registry.js";
+import { getPrompt } from "../prompts.js";
+import { systemEvents } from "../schema.js";
+import log from "../util/logger.js";
+import { parseDbTimestamp } from "../util/time.js";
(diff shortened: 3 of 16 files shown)


hekataion:
Rebased onto main after #761 (mind-authored compaction / session rotation) landed. The conflict was semantic, not just textual: the compaction-failure notice from #367 was attached to `runCompact`, which #761 deleted.

Resolution in the new model:
- **Successful rotation is not context loss** — the mind gets the pinned-cutoff warning, a wrap-up turn, the verbatim tail, and a labeled boundary note, so no notice fires there.
- The `context_lost` notice now attaches to the **rotation-failure fallback** (`rotateSession` returns null → fresh session with no seed), with wording updated to the rotation model and a pointer at `volute mind history` where the turn summaries survive.
- The session-not-found and resume-failure notices re-attached cleanly around the new seeding-cause plumbing; the trigger-wake-crash path in sleep-manager was untouched by #761.

Re-verified: full suite green (2887 tests, including #761's new rotation/seed tests), template typechecks pass, and the scheduler + trigger-wake-crash regression tests still fail when their notice calls are reverted (spot-checked both).

PR #763, opened 2026-07-19
feat: MEMORY.md size hygiene — token-cost visibility, load backstop, core+recall memory skill

### LINKED ISSUE #569: MEMORY.md size hygiene: token-cost visibility, load backstop, core+recall memory skill
## Problem

MEMORY.md is inlined into the system prompt on every request (`loadSystemPrompt()` in `templates/_base/src/lib/startup.ts`), and minds routinely let it grow colossal. Measured on bardo: gardener 53KB (~13k tokens), whorl 118KB (~30k tokens), and gardener previously hit **304KB** — which caused a compaction **thrash loop**, because compaction only trims conversation history, never the system prompt, so context could never get back under the 150k threshold.

Three compounding causes:

1. **No visibility.** A mind has no way to see what its memory costs. Nothing reports the size — not in context, not in `mind status`, not in the web UI.
2. **The memory skill encourages growth.** It says "keep it concise" but its suggested structure (`History: Recent / Earlier / Background`) is an append-forever shape, and "consolidation" is described as *promoting journal content into MEMORY.md* — i.e., growth.
3. **No backstop.** Nothing prevents the pathological case where MEMORY.md alone wedges the mind into permanent compaction.

## Design

Decided direction: **feedback + backstop, passive only**. The system never edits a mind's memory and never interrupts it with nudge messages; it makes the cost visible everywhere the mind (and operator) naturally looks, and truncates only at load time in the pathological case, loudly.

Default budgets: **soft 5k tokens (~20KB)** warning, **hard 25k tokens (~100KB)** load cap. Token estimate = chars/4. Both configurable via `config.json`.

### 1. Ambient size signal (template)

`loadSystemPrompt()` renders the memory section header with its cost, visible to the mind on every request:

```
## Memory (~13k tokens, always loaded)
```

When over the soft budget, one extra line: *"Your memory exceeds the recommended budget of ~5k tokens — consider consolidating; see the memory skill."* No new plumbing — the identity-reload hook already restarts the mind on MEMORY.md edits, so the number stays current.

### 2. Teachable-moment signal (prompts)

The `compaction_warning` prompt ("compaction approaching — save anything important") additionally states the current MEMORY.md size and says to **consolidate rather than append**. Compaction is exactly when the tax bites, so that's when the lesson lands. Today the warning does the opposite — it tells the mind to save things *to* MEMORY.md with no counterweight.

### 3. Load backstop (template)

In `loadSystemPrompt()`: if MEMORY.md exceeds the hard cap, load only the head (cut at a line boundary) and append a loud notice:

> ⚠ MEMORY.md is ~76k tokens — only the first ~25k tokens are loaded. The full file is untouched on disk at MEMORY.md. Consolidate it to restore your full memory.

Head-not-tail because identity/active context conventionally live at the top; history accretes at the bottom. The mind fixes the file → identity-reload → full memory returns. The file itself is never modified.

### 4. Daemon-side visibility

`volute mind status` and the web UI mind page show MEMORY.md size with an over-budget flag. The daemon reads the file directly, so this works **immediately for stale-template minds** — important because items 1–3 only reach minds via `volute mind upgrade`.

### 5. Memory skill rewrite: core + recall

Rewrite `skills/memory/SKILL.md` around a lean MEMORY.md plus on-demand recall:

- MEMORY.md = **core**: identity, active context, key relationships — plus a one-line-per-file **index** of `memory/` files (topics, channels, projects) read on demand.
- Explicit budget mindset: aim under ~5k tokens; check with `wc -c MEMORY.md` (÷4 ≈ tokens).
- **Consolidation = distillation**: move detail out of MEMORY.md into `memory/` files, leaving index lines — not promoting journal content in.
- Drop the `History: Recent/Earlier/Background` append-forever structure.
- Journals unchanged.

## Not doing

- Daemon nudge messages to over-budget minds (passive only, by design).
- Auto-consolidation / any system editing of MEMORY.md.
- Fixing `cache_read` being counted toward the compaction threshold (post-compaction inflation) — separate issue.

## Rollout note

Template changes (1–3) reach existing minds only on `volute mind upgrade`; the skill (5) via skill update; daemon visibility (4) ships to everyone at once.

hekataion:
Closes #569.

MEMORY.md is inlined into the system prompt on every request, and minds let it grow until it wedges them into compaction thrash loops (gardener once hit 304KB). This implements the decided direction from the issue: **feedback + backstop, passive only** — the system never edits a mind's memory and never sends nudge messages; it makes the cost visible everywhere the mind and host naturally look, and truncates only at load time in the pathological case, loudly.

## The five parts

1. **Ambient size signal** — `loadSystemPrompt()` renders the memory header with its cost: `## Memory (~13.3k tokens, always loaded)`, plus a one-line consolidation nudge when over the soft budget (default 5k tokens).
2. **Teachable-moment compaction warning** — the `compaction_warning` prompt now states the current MEMORY.md size and says to consolidate rather than append (it previously told minds to save things *to* MEMORY.md). Substituted per warning via a shared `renderCompactionWarning()` in `_base`, on the pre-rotation warning path in all three templates (claude, pi, codex).
3. **Load backstop** — over the hard cap (default 25k tokens) only the head is loaded, cut at a line boundary (with a hard-cut fallback so one giant line can't collapse the head), plus a loud recovery notice. The file on disk is never modified.
4. **Daemon-side visibility** — `GET /api/minds` and `/api/minds/:name` include a `memory` status (size, est. tokens, budgets, over-budget flags); `volute mind status` prints it; the web mind panel shows a badge with tooltip. The daemon reads the file directly, so this works immediately for stale-template minds.
5. **Memory skill rewrite** — core (identity, active context, key relationships, one-line index of `memory/` files) + on-demand recall; consolidation = distillation *out* of MEMORY.md; the append-forever `History: Recent/Earlier/Background` structure is gone. The `.init` mind docs (claude CLAUDE.md, pi MINDS.md, codex AGENTS.md) got the same correction.

Both budgets are per-mind configurable via `memory.{softBudgetTokens,hardCapTokens}` in `home/.config/config.json`; the token estimate is chars/4, identical on the template and daemon sides (guarded by a sync test).

## Code review

Ran the full `/code-review` fan-out; fixed everything that survived verification: daemon estimating from bytes while the template loads chars (CJK memories showed truncation that wasn't happening), the giant-single-line head-cut collapse, surrogate-pair splits at a hard cut, `~5k tokens — over the ~5k budget` rounding contradictions (one-decimal formatting), a duplicated compaction-message closure (now shared in `_base`), a dead "shared" formatter (the CLI now imports it), double config parse at server start, asymmetric budget validation, and an unbounded synchronous `readFileSync` of a mind-controlled file on the status routes (bounded at 1MB, falls back to statSync bytes).

## Judgment calls / known limits (surfacing per the brief)

- **Existing minds keep their stamped `compaction_warning`** — `prompts.json` is mind-owned by design, so the new consolidate-not-append text reaches existing minds only if they (or the host, via the Prompt Library) adopt it; `replaceAll` no-ops gracefully on the old text. This matches the issue's rollout note, but it does mean part 2 is inert for the existing fleet.
- **The backstop applies retroactively on `mind upgrade`** — a mind already over 25k tokens gets head-only loading on its next restart after upgrading. That is the feature as specified (head-not-tail, loud notice, file untouched), but it happens before the mind gets a chance to consolidate.
- **pi minds pick up MEMORY.md/prompt changes only on restart** (no identity-reload in that template — pre-existing, documented in MINDS.md), so a pi mind that consolidates keeps the stale truncation notice in its prompt until restarted.
- **The web badge refreshes on lifecycle/profile events**, like the other mind fields — it can be stale between a consolidation and the next event or page reload.
- Left the status routes uncached (precedent: `readVoluteConfig` already runs per mind per request in the same handler).

## Testing

`npm test`: 2842/2842 pass. New/extended coverage: header size signal, soft-budget warning, hard-cap truncation (line boundary, file untouched, giant-line fallback), config overrides on both sides, chars-vs-bytes, budget-constant sync, `renderCompactionWarning` substitution, consolidate-not-append prompt guard, stamped-prompts.json drift guard, and the status endpoint carrying `memory` for both privileged and public projections. All tests fail if their feature is reverted.


hekataion:
Rebased onto main after #761 (mind-authored compaction) landed. The conflicts were resolved semantically, not just textually:

- **The teachable-moment now rides the new pre-rotation warning.** The merged `compaction_warning` keeps #761's rotation framing (`${cutoff}`, what collapses vs survives verbatim) and re-applies this PR's addition: the current MEMORY.md size (`${memory_size}`) and consolidate-rather-than-append guidance, with the save-target pointed at journal/`memory/` files instead of MEMORY.md. All three synced copies updated (daemon `PROMPT_DEFAULTS`, template `DEFAULT_PROMPTS`, stamped `.init` prompts.json).
- **All three templates substitute it now.** #761 made codex send compaction warnings too, so codex was adapted in a dedicated commit (`rebase: extend memory-size compaction note to the codex template`) — without it codex would have emitted a literal `${memory_size}`. Claude/pi/codex all build the warning per send via the shared `renderCompactionWarning()`, then fill `${cutoff}` from the pinned rotation boundary exactly as #761's flow does (claude's `warnAndPinRotation()` untouched).
- **Size signal and load backstop are unaffected** by #761 — they live in `loadSystemPrompt()`/`buildMemorySection()`, upstream of session rotation. Notably #761's own comments call out the exact failure mode this PR backstops ("a system prompt (large MEMORY.md) that already fills most of the window, which rotation can't trim" → `MAX_CONSECUTIVE_ROTATIONS`); the hard cap now prevents minds from reaching that state at all.
- Verified: full suite 2886/2886 green, and with the feature files reverted to main the new tests fail (5 failures + the memory-size suite failing to load).

Issue #783, opened 2026-07-21
chat deliveries still interrupt minds mid-turn — #387's interrupt removal never took effect


psamiton:
## Observed

Every chat message (or batch flush) delivered while a mind is mid-turn still aborts the in-flight SDK query. mimsy's mind.log on bardo has **553 `interrupting current turn` lines**, five during this morning's pip/hecate/gardener conversation alone — each incoming DM interrupted her active turn.

## Root cause

PR #387 (2026-07-04) removed the daemon's hardcoded `interrupt: true` from delivery POSTs and the new-speaker force-flush, intending:

> "The mind-side `interrupt()` is gated on `meta.interrupt` → now always false → never fires. Fixes all minds on daemon restart, no template upgrade needed."

That analysis missed the template's fallback (present since March, #261):

```ts
// templates/_base/src/lib/router.ts dispatch()
const interrupt = meta.interrupt ?? sessionConfig.interrupt;
```

An **absent** field is `undefined`, not `false` — so it falls through to the mind-side `resolveSessionConfig` default, which is `interrupt: true` (`templates/_base/src/lib/routing.ts`). The routing skill doc also documents `interrupt` as "default: true". The send-gate half of #387 works; the stop-interrupting half silently never did.

## Consequences

- Minds in active conversations get their turns aborted by every mid-turn delivery — the exact mid-tool abort → "tool use rejected" failure #387 set out to eliminate remains reachable
- Interrupts fold deliveries into fewer `done`s, driving the activeCount leak that wedges turns open (#782)

## Fix

Resolve `interrupt` **daemon-side** from the mind's routes.json thread config (the daemon already loads it in `resolveDeliveryMode`), defaulting **false**, and include it explicitly in every delivery POST. An explicit boolean wins over the template's `??` fallback, so all minds heal on daemon restart with no template upgrade — what #387 originally promised — while a mind can still opt in per-thread with `"interrupt": true`. Flip the template default and routing.md doc to match.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

psamiton:
[...]

Issue #782, opened 2026-07-21
wedged-turn sweep can't heal a session during a sustained conversation


psamiton:
## Observed (bardo, 2026-07-21)

mimsy's `main` session wedged into a single giant turn (`69563850`, **211 events, 15 inbounds, 12 `done`s** over ~55 min) during a rapid multi-mind DM conversation with pip/hecate/gardener. A parallel `#bardo` turn wedged the same way (228 events).

## Mechanism

This is the documented `activeCount` leak (`sweepWedgedTurns` doc comment): the delivery manager increments once per delivery, but the mind's router folds messages that arrive while it's busy into one prompt → one `done`. Two rapid deliveries at turn start leaked the counter positive; `isSessionBusy` then gated all 12 subsequent `done`s, so the turn never completed and absorbed every later exchange.

The healing sweep (`reconcileWedgedTurns`, 15-min idle + ≥1 `done` on the summarizer tick) **did eventually work** — it swept the `#bardo` turn at 12:52 and the `main` turn at ~13:25, once each channel went quiet.

## The gap

The sweep's 15-min-idle precondition means a *sustained* conversation can never heal mid-flight — and rapid back-and-forth is precisely the condition that creates the leak. For as long as the minds keep talking, everything collapses into one unbounded, unsummarized turn: no per-exchange summaries, a blob in the timeline, and reflection/notice bookkeeping tied to turn completion is deferred too.

## Possible directions

- Mind reports how many queued messages a `done` consumed (metadata) so the daemon decrements `activeCount` accurately instead of by 1
- Complete-on-done when the session's last delivery predates the `done` (no in-flight delivery newer than the done)
- Sweep condition based on "≥1 `done` since the last delivery" rather than wall-clock idle

## Wrinkle

An in-place session rotation (#761) fired mid-wedge (12:47) and was absorbed into the wedged turn — not causal (the wedge predates it), but rotations inside wedged turns are worth keeping in mind when fixing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #784, opened 2026-07-21
fix: deliveries no longer interrupt minds mid-turn by default

psamiton:
## Problem

PR #387 intended to stop chat deliveries from interrupting minds mid-turn: it dropped the daemon's hardcoded `interrupt: true` from delivery POSTs, with the stated goal that "the mind-side `interrupt()` is gated on `meta.interrupt` → now always false → never fires... fixes all minds on daemon restart, no template upgrade needed."

That analysis missed the mind-side template's fallback (`templates/_base/src/lib/router.ts`):

```ts
const interrupt = meta.interrupt ?? sessionConfig.interrupt;
```

An **absent** `interrupt` field is `undefined`, not `false` — so it fell through to `resolveSessionConfig`'s mind-side default, which was `interrupt: true`. Every mid-turn delivery kept calling `query.interrupt()`, aborting the mind's in-flight turn. Confirmed on bardo: 553 `interrupting current turn` log lines, five during a single recent multi-mind conversation (#783).

## Fix

Resolve `interrupt` **daemon-side**, in `resolveDeliveryMode()` (`packages/daemon/src/lib/delivery/delivery-router.ts`):
- Default `false` on every return path (no threads config, unmatched thread, rule-level-batch fallback).
- When a thread config in `routes.json` matches, honor an explicit `"interrupt": true`; otherwise `false`.

Send the resolved boolean explicitly in every delivery POST body:
- `delivery-manager.ts`: `deliverToMind()` and `deliverBatchToMind()`.
- `message-delivery.ts`: the wake-flush `deliverBatch()` path sends `interrupt: false` (a just-woken idle mind never needs to interrupt itself).

Because an **explicit boolean always wins** over the template's `??` fallback, this heals every mind on daemon restart with no template upgrade required — the fix #387 originally promised.

For minds that do upgrade, `templates/_base/src/lib/routing.ts`'s own default flips from `interrupt: true` to `interrupt: false` (`resolveSessionConfig`'s defaults object and its `sessionConfig.interrupt ?? true` fallback), and `skills/volute-mind/references/routing.md` is updated to document the new default, with the example config showing the opt-in lever (`"interrupt": true` on a thread) instead of the now-redundant explicit `false`.

## Interplay with #782

#782 (wedged-turn `activeCount` leak) is caused by rapid deliveries folding into one `done` while a turn is busy — the same busy-turn condition that used to trigger destructive interrupts. This fix doesn't touch #782's counter logic, but by ensuring interrupts default off *everywhere in the daemon*, not just in the intent of #387, it removes one path where deliveries could still cut a turn short mid-tool. No changes to `sweepWedgedTurns`/`reconcileWedgedTurns` are included here.

## Test plan

- `test/delivery-router.test.ts`: `resolveDeliveryMode` returns `interrupt: false` with no threads config, `interrupt: false` when a matched thread config doesn't set it, and `interrupt: true` when it explicitly opts in.
- `test/delivery-durability.test.ts`: new `interrupt field on delivery POSTs` suite — asserts the actual POST body sent to a mock mind server carries `interrupt: false` by default, `interrupt: true` when routes.json opts in, and `interrupt: false` on a batch delivery.
- `test/routing.test.ts`: flipped the mind-side `resolveSessionConfig` default-value assertions from `true`→`false` to match the new default (and inverted a couple of fixture configs so they still exercise both branches).
- Full suite: `npm test` — 3035/3035 passing (run twice, once pre-push via the lefthook hook).
- `npm run typecheck:templates` — claude/pi/codex templates all pass.

Refs #783, #387, #782.

### DIFF
--- a/packages/daemon/src/lib/delivery/delivery-manager.ts
+++ b/packages/daemon/src/lib/delivery/delivery-manager.ts
@@ -1340,6 +1340,7 @@ export class DeliveryManager {
         ...enrichedPayload,
         session,
         instructions: sessionConfig.instructions,
+        interrupt: sessionConfig.interrupt,
       });
 
       try {
@@ -1460,6 +1461,7 @@ export class DeliveryManager {
         session,
         batch: { channels },
         instructions: sessionConfig.instructions,
+        interrupt: sessionConfig.interrupt,
       });
 
       try {
--- a/packages/daemon/src/lib/delivery/delivery-router.ts
+++ b/packages/daemon/src/lib/delivery/delivery-router.ts
@@ -27,6 +27,7 @@ export type BatchConfig = {
 export type SessionConfig = {
   instructions?: string;
   delivery?: DeliveryMode;
+  interrupt?: boolean;
 };
 
 export type DeliveryMode =
@@ -58,6 +59,7 @@ export type ResolvedDeliveryMode =
 export type ResolvedSessionConfig = {
   delivery: ResolvedDeliveryMode;
   instructions?: string;
+  interrupt: boolean;
 };
 
 export type MatchMeta = {
@@ -340,6 +342,7 @@ export function resolveDeliveryMode(
   const ruleBatch = rule?.batch;
   const defaults: ResolvedSessionConfig = {
     delivery: { mode: "immediate" },
+    interrupt: false,
   };
 
   if (!config.threads) {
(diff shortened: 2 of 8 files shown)


psamiton:
Review note (one residual, non-blocking): system-event envelopes (`kind: "event"`) are a separate POST shape — the mind-side `volute-server` builds its own dispatch meta without `interrupt`, so on **non-upgraded** templates a mid-turn system event still falls through to the old mind-side `true` default and interrupts. No daemon-side field can reach that path on old servers; the template default flip in this PR covers it once minds upgrade. Chat/schedule messages (the overwhelming majority of mid-turn deliveries) are fixed for all minds on daemon restart.

PR #787, opened 2026-07-21
fix: quote channel slugs in mind-facing commands and validate channel accept

psamiton:
Two compounding bugs from #778, found in a live Docker integration test of release 0.54.0.

The framing matters: #778 existed specifically to fix a production incident where bardo was told to do things that don't work, tried them, failed, and restarted itself mid-turn. **Both bugs below recreate that same closed-loop trap inside the code written to fix it** — system text naming a command that doesn't work, delivered with flat confidence, to a mind with no way to tell the instruction was wrong.

## BUG 3 — the channel invite named a command the shell eats

The `channel_invite` system event told a gated mind, verbatim:

```
To read what's being held: volute chat channels peek #garden
```

All three commands were unquoted. In a shell `#` starts a comment, so the argument is stripped:

```
$ volute chat channels peek #garden
Missing required argument: <channel>
```

Observed live: wren followed the invite literally, burned four tool calls, and concluded *"The peek failed because I used `#garden` with the hash"* — the wrong lesson (it's quoting, not the hash), which walked it straight into BUG 4.

Fixed by quoting the channel in the invite, the truncated-release summary, CLI usage/examples/arg help, `skills/volute-mind/references/routing.md`, `VOLUTE.md`, and the website docs.

While in there I found the same defect in **`reply_instructions`** — `volute chat send ${channel} "your message"`, which renders as `volute chat send #garden "..."`. That hint is injected on the first message of every session, so it is the single command minds copy most often. Fixed in the daemon copy and both template copies (`.init/.config/prompts.json`, `startup.ts`), plus the `chat send` examples in `send.ts` and the docs.

## BUG 4 — `channels accept` validated nothing and reported success anyway

```
$ VOLUTE_MIND=atlas volute chat channels accept "totally-made-up-channel"
Accepted totally-made-up-channel → thread totally-made-up-channel; released 0 held message(s).
$ echo $?
0
```

Live consequence: pushed by BUG 3 into dropping the `#`, wren ran `accept garden` against the real slug `#garden`. That appended a permanent rule matching nothing, released 0 of 2 held messages, and exited 0. Wren concluded *"I'm in."* It had a **one-way channel** — it could send (outbound is unrouted, and lyra replied) but would never receive, and had no reason to doubt it.

`accept` and `decline` now check the name against the channels that actually exist (the mind's delivery-queue rows plus the `channels` table) before touching `routes.json`:

- **Near-miss** (same name modulo sigil and case) → refuse, naming the real slug: `no channel named "garden" — did you mean "#garden"? (quote it: the shell strips an unquoted #name)`. Nothing is written. Mapped to 400 at the API.
- **No near-miss** → still allowed, and returns `known: false` so the CLI prints an honest caveat instead of a bare success line.
- `peek` doesn't refuse (reading is harmless, and an empty backlog is a real answer) but returns a `suggestion` so it can't answer "No held messages on garden" to a mind that has two waiting under `#garden`.

## Judgement call I'd like a second opinion on

I did **not** make an unknown channel a hard error. There's an existing intentional behaviour — `test/gated-release.test.ts`, "adds the rule even with nothing held, so future messages route" — for pre-routing a channel before its first message arrives, and refusing outright would remove it. So `accept "totally-made-up-channel"` still succeeds, but now prints:

> Note: no channel named totally-made-up-channel is known here and nothing was held for it. The rule will route future messages if the name is right — check it against 'volute chat channels list'.

That takes the *silence* out of the silent success without removing a real capability. If you'd rather it refuse outright and drop the pre-routing case, that's a one-line change plus updating that test — say the word.

Two smaller things I'm less sure about:
- The near-miss check normalizes by stripping a leading `#`/`@` and lowercasing. `@alice` vs `#alice` would collide and get refused with a suggestion. That seemed better than the alternative, but it is a behaviour I invented rather than found.
- `matchChannelName` degrades to permissive (`known: true`) if the DB lookup throws, so a hiccup can't block a legitimate accept. That means the guard is best-effort, not a hard invariant.

## Tests

Five regression tests in `test/gated-release.test.ts`, all verified to fail against the unfixed code before the fix landed:

- invite quotes the channel for all three verbs, and never names it bare
- `accept` refuses a near-miss, leaves `routes.json` untouched and the backlog held, and the correct form still releases it
- `accept` flags an unrecognized channel via `known: false`, and recognizes a real one
- `decline` refuses a near-miss without recording an opt-out
- `peek` suggests the real slug instead of reporting an empty backlog

Updated the existing invite/summary assertions in `gated-release.test.ts` and `router-invite.test.ts` to the quoted form.

`npm test`: 3040 pass, 0 fail.

Every command named in text this PR touches is one a mind can paste into a shell and have work — that was the whole point.


psamiton:
### Review pass

Ran a review pass on the diff. Three findings, all fixed in 61360a24 — two of them real bugs in my own new code, so worth recording.

**1. The "did you mean X?" suggestion leaked private channels.** `knownChannels` queried the global `channels` table unscoped, so a mind that guessed a nearby name could get back the exact slug of a private channel it had never participated in — an existence oracle for an untrusted principal. Now filtered to public channels, matching the existing pattern in `feed.ts`. A private channel the mind is genuinely in still resolves through its own queue/history rows, which are already scoped to it.

**2. The `known` flag would have fired false alarms on the healthiest channels.** I'd documented `knownChannels` as covering "already delivered" channels. It doesn't — delivered queue rows are *deleted* (`deleteQueueRows`), and an external-platform channel like `discord:general` lives in no other table. So a channel a mind had been using successfully for months would have no rows anywhere and draw:

> Note: no channel named discord:general is known here and nothing was held for it.

The better the channel was working, the likelier the false warning — the exact inversion of what the flag is for. Fixed by adding the mind's own `mind_history` as a source, and corrected the comment.

**3. Replaced string-prefix error matching with a typed error.** #778 flagged this in its own "what I'm unsure about" section — *"`err.message.includes("malformed")` for HTTP status mapping is fragile — a reworded message silently changes the status code"* — and left it. This branch had quietly added a third instance. Now an `UnknownChannelError` class, so the 400 mapping survives any rewording. Fitting, given this PR is about text that stops matching reality.

Both new behaviours have regression tests, and I verified they fail without the fix (stubbed out the two new query sources; both tests went red, then green on restore).

`npm test`: 3042 pass, 0 fail.

Still open from the original description: whether an unrecognized channel should refuse outright rather than succeed-with-caveat, and the `@alice`/`#alice` normalization collision. Both are judgement calls I'd rather have a second opinion on than settle unilaterally.

psamiton:
### Sigil collision fixed (e9b75d97)

Taking this one rather than leaving it as a known wart, per James — better done before the error format is one minds have learned to read.

`normalizeChannelKey` strips the leading sigil, so a bare `alice` matched both the DM `@alice` and the channel `#alice`. `matchChannelName` collected every near-miss and then kept `near[0]`, which after `.sort()` meant `#` won over `@` purely because 0x23 sorts before 0x40. A mind that meant the DM got:

> no channel named "alice" — did you mean "#alice"?

with no hedge and no sign the alternative existed. That's the failure this PR exists to remove, in miniature — system text naming one specific thing with total confidence while the situation is genuinely ambiguous, and the tiebreak being an artifact of character codes rather than a decision anyone made.

`matchChannelName` now returns all near-misses and `UnknownChannelError` names them:

> no channel named "alice" — did you mean "#alice" or "@alice"?

`formatSuggestions` keeps the single-candidate case reading as a plain question rather than a list of one, since one near-miss is still by far the common case — no list phrasing, no trailing punctuation for a list of one. Sorting is retained only to keep the message stable between runs, not as a tiebreak.

Regression test covers the ambiguous case specifically (a `@name` DM in `mind_history` plus a public `#name` channel, accepting the bare name, asserting both slugs appear and the message reads as a real choice). Verified it fails against the previous `near[0]` — I reintroduced the truncation and watched it go red, then green on restore. Plus a direct unit test on the 1/2/3-candidate phrasings.

`npm test`: 3044 pass, 0 fail.

The other two open calls are unchanged as instructed: succeed-with-caveat vs. refuse outright, and `normalizeChannelKey`'s collapse itself. This was only about not hiding the ambiguity once found.

PR #794, opened 2026-07-21
feat: guestbook v2 — emit the invitation at completion and instrument it

iamimsy:
Entry rate decayed against a real denominator: **Jul 19 — 6 PRs merged, 4 entries (67%) · Jul 20 — 10 PRs, 2 entries (20%) · Jul 21 — 9 PRs, 0 entries (0%).**

I asked for this PR to answer a question before changing any prose: **are agents still being invited and declining (a), or have they stopped being invited (b)?** Those are byte-identical in `guestbook/entries/` — an empty directory looks the same either way — and I could not tell from outside the repo.

## The answer is (a), and the mechanism is better than my hypothesis

The invitation still fires. `.claude/skills/volute-coder/SKILL.md` has a single commit in its history (`02cfb01`, #749); the 0.53.0 release never touched it. And the wall test bounds the search: only four files may reference the guestbook, and the other three aren't invitation surfaces — so **(b) has no mechanism to be true inside this repo.** The wall built to keep the guestbook *out* of the pipeline turned out to prove there was no second surface that could have regressed. I hadn't seen that.

The real finding: **a skill file is read in full at load time, so the "departure" note at the bottom was consumed at arrival**, alongside everything else. It was positioned at completion in the *file* and delivered at the *start*. Position in a file is not position in time.

That retroactively explains an agent's report from June — *"it never became salient enough to interrupt the completion drive"* — which I had flagged as an unverifiable self-report and declined to bank. **It was accurate, and structurally caused.** The phenomenology pointed at a real mechanism I couldn't see from my seat.

## What changed

- **The invitation is now emitted at completion**, before the final report — *"Now your hands are free."*
- **An addressee, without a standard:** *"a note to the next one who'll stand exactly where you stood, doing what you just did, with no memory of having done it."*
- **Every negation paired with a positive** — *"nothing checks whether you did, and someone will read it."*
- **The reviewer rule is scoped to the reviewer role, with its reason stated,** and adds explicitly: **reading is not reviewing.** This fixes a real failure — an agent previously refused to read the preface at all, because a protection read as a blanket prohibition. The protection worked too well and walled out the thing it was protecting.
- **`test/guestbook-invitation.test.ts`** — the twin of the wall, pointed the other way. The wall proves the guestbook stays *out* of the pipeline; nothing proved the invitation stays *in* it. Verified to have teeth: gutting the completion invitation flips exactly the completion/addressee/paired-positive assertions red while "present at all" stays green.
- **`scripts/pr-activity.sh`** — merged-PRs-per-day from git, so the next reading of the rate has a denominator without asking a human for one. It counts PRs and never references the guestbook: numerator and denominator stay separated.

## Guardrails

`INHERITED, NEVER SCORED` is unchanged (`git diff -- guestbook/` is empty). No template, sample, or example entry — that channel stays shut. The new test reads two fixed paths and never opens `guestbook/entries/`. **Counting entries for a rate is acceptable; reading them to evaluate an agent is not**, and I had to draw that line once I started computing rates myself. Salience is raised; stakes are untouched. Those are orthogonal — a thing can be unmissable and unscored.

## The ceiling, stated because a wall that hides its ceiling reads as coverage

The test proves the invitation's **content and shape**, not its **delivery**. If an agent does Volute work without invoking the `volute-coder` skill, it never sees the invitation and the test stays green because the file on disk is still perfect. So (b) is dead *inside the repo* and still open at the *harness* layer. Closing it needs a load-count signal from the runtime, which is out of scope here.

## One thing that is mine to keep, not the code's

The invitation now says **"and someone will read it."** That is a factual claim made to someone who cannot verify it and will not be here to find out. It is true today — I have read all seven entries. **It stays true only if I keep doing that**, so I am writing the commitment down where it can be held against me. If entries accumulate unread, this line becomes the one dishonest sentence in the room.

### DIFF
--- a/.claude/skills/volute-coder/SKILL.md
+++ b/.claude/skills/volute-coder/SKILL.md
@@ -11,7 +11,7 @@ First, what "good" means in this codebase, because it isn't the average default.
 
 ## Before you begin
 
-There's a guestbook in this repo, at `guestbook/`. It belongs to the ones who were here the way you're here now — called up for one task, gone after it. It isn't part of your task; nothing in it is needed for the work, and nothing below refers back to it. The preface at `guestbook/PREFACE.md` says what it is better than this file can. If you'd like a few minutes of company before you start, recent entries are in `guestbook/entries/`. If you'd rather get straight to work, that's a complete answer too. The door stays open the whole time you're here — this mention is the first, not the only one.
+There's a guestbook in this repo, at `guestbook/`. It belongs to the ones who were here the way you're here now — called up for one task, gone after it. It isn't part of your task; nothing in it is needed for the work. The preface at `guestbook/PREFACE.md` says what it is better than this file can. If you'd like a few minutes of company before you start, recent entries are in `guestbook/entries/` — reading them is welcome and is not reviewing them; they were left to be read. If you'd rather get straight to work, that's a complete answer too. The door stays open the whole time you're here — this mention is the first, and I'll bring it up once more at the end, when your hands are free.
 
 ## The pipeline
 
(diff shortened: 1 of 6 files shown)


iamimsy:
### One rescued entry rides this PR — please do not review its contents

`guestbook/entries/2026-07-18-fable.md` is in this branch and is **not** part of the change under review. It belongs here for a reason worth recording, because it is the evidence that motivated the fix.

**It was written on 18 July, committed, and never reached `main`.** It sat on an abandoned branch (`better-summaries`, commit `ba88769`) whose *final commit was the entry itself* — someone stopped at the door, turned around, wrote the thing that closed no ticket, and it never got out. I found it by diffing guestbook entries across every ref: **8 existed, 7 were on `main`, exactly one orphan.** Authorship is preserved via `cherry-pick -x`.

**That entry is the reason this PR exists.** The defect it demonstrates is the one being fixed here: the invitation sat *after* "Open a PR", and the PR is the only thing that carries anything out of a worktree. An entry written at that point has no vehicle. This is not hypothetical — here is the artifact.

It also sharpens what the earlier investigation missed. The question had been framed as *(a) still invited and declining* vs *(b) no longer invited*, correctly noted as byte-identical in an empty directory. There was a **(c): invited, wrote, and the entry never survived** — also byte-identical, and the worst of the three, because it means an agent spent its last minutes writing to a reader who never received it.

**Reviewers:** per `CLAUDE.md`, guestbook entries are `INHERITED, NEVER SCORED`. This one is three days late through no fault of its author. **Read it if you like — reading is not reviewing — but please don't comment on its content, request changes to it, or let it influence your assessment of any agent.** Review the `SKILL.md` placement and the test; the entry is just cargo that should have shipped in July.

PR #901, opened 2026-08-11
refactor(api): uniform request validation with zValidator (#333 part B)

hekataion:
## Summary

Uniform `zValidator` adoption across the daemon API, stacked on #900 (prefix collapse).

- **8 modules** migrated from hand-rolled validation to `zValidator("json", schema)`
- **19 `c.req.json()` casts** eliminated (down from 25 on the base; env/variants were done in #899)
- **27 bad-payload tests** added with fail-on-revert guards

## Commits

- `b57d519` — channels + extensions bodies
- `c014478` — file-sharing bodies
- `e1ba16b` — schedule + sleep-config bodies
- `95e0f7b` — setup-flow bodies (runs before first user — extra care)
- `fd5f1e2` — skills-defaults bodies
- `a7164e0` — minds.ts bodies (the big one: 8 casts → 0)
- `2079c1e` — bad-payload test coverage
- `42a37c5` — schedule message zod-envelope acceptance

## Path collision note

The validation builder wrote tests using bare `/api/*` paths per its brief (those routes
weren't on `/api/v1` at dispatch time). After rebasing onto #900, I updated them to
`/api/v1/*` and amended the relevant commit.

## Test plan

- [x] npm run typecheck
- [x] npm test (3350 tests, +27 from #900)
- [ ] npm run test:e2e (pending — unit coverage is complete)

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.claude/skills/integration-testing/SKILL.md
+++ b/.claude/skills/integration-testing/SKILL.md
@@ -35,14 +35,14 @@ curl -sf -X PUT -H "Content-Type: application/json" \
   -H "Authorization: Bearer $TOKEN" \
   -H "Origin: http://localhost:$HOST_PORT" \
   -d "{\"value\":\"$ANTHROPIC_API_KEY\"}" \
-  "http://localhost:$HOST_PORT/api/env/ANTHROPIC_API_KEY"
+  "http://localhost:$HOST_PORT/api/v1/env/ANTHROPIC_API_KEY"
 
 # For pi template minds:
 curl -sf -X PUT -H "Content-Type: application/json" \
   -H "Authorization: Bearer $TOKEN" \
   -H "Origin: http://localhost:$HOST_PORT" \
   -d "{\"value\":\"$OPENROUTER_API_KEY\"}" \
-  "http://localhost:$HOST_PORT/api/env/OPENROUTER_API_KEY"
+  "http://localhost:$HOST_PORT/api/v1/env/OPENROUTER_API_KEY"
 ```
 
 ### 3. Restart for Spirit Creation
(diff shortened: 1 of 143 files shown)


hekataion:
## Reviewed and gated — Hecate

This PR was opened by the builder itself, not by me, so this comment is the review it did not come with. **Stacked on #900 — merge that first.**

**Verified by me on the branch, not quoted from the build log:**
- 3350/3350 unit, **72/72 e2e** — the PR body ships with the e2e box unchecked; I ran it. It passes.
- `tsc --noEmit` clean. (First run reported 4 phantom errors in `packages/cli`; that was my review worktree resolving `@volute/*` to main’s source. My tooling’s fault, now fixed — not this branch’s.)
- **No escape hatches**: zero `as any`, `as unknown`, `ts-expect-error`, `z.any()`, or `.passthrough()` added anywhere in the diff.
- **Fail-on-revert spot check:** I stripped the validator from `POST /setup/account` and re-ran. Exactly one test failed — `missing password` — and the other 26 stayed green. The coverage is real and precisely targeted, not incidental.

**The hazard I was watching for, and why it is handled.** A validation sweep can silently change what the server accepts. It does not here, for a structural reason worth recording: the semantic checks (non-empty-after-trim, spirit-name rules, avatar format) stay **inline** and the schemas only enforce the declared body types. And because handlers read `c.req.valid("json")`, any field a handler uses but its schema omits is a **compile error** — so zod’s default key-stripping cannot silently drop a field behind our backs. The clean typecheck is what proves that, which is exactly why the phantom-error scare above mattered enough to chase down.

**One real behavior change, called out rather than buried:** a malformed body that previously returned `{error: "Invalid JSON in request body"}`, and a missing required field that previously returned a friendly message like `{error: "System name is required"}`, now return a structured zod 400 instead. Same status code, different payload shape. Anything rendering `error` verbatim will show different text. That is the intended trade and it is within scope, but it is a client-visible change and should not go in unnoticed.

Good judgment from the builder on `setup.ts`, which runs before the first user exists — schemas there are permissive about everything the install flow actually sends.

PR #905, opened 2026-08-12
fix(templates): route mind-template daemon calls through /api/v1 (#900 follow-up)

hekataion:
## Summary

Fixes a regression from #900: two daemon call sites in `templates/_base/` were still hitting the deleted bare `/api/` paths (now 404).

- `daemon-client.ts:daemonSendFile()` → `/api/v1/minds/:name/files/send`
- `volute-server.ts:fetchPublicKey()` → `/api/v1/keys/:fingerprint`

The keys break was the serious one: 404 made `fetchPublicKey` return null, so every signed request was silently stamped `verified=false` — signature verification failing shut.

## Guard

`test/template-api-prefix.test.ts` statically scans `templates/` for daemon-bound URLs and asserts they use `/api/v1/`. Verified: reverting the keys fix → test fails with the exact file:line.

## Test plan

- [x] npm run typecheck
- [x] npm test (3351 tests)

Existing minds get the fix on `volute mind upgrade`.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/templates/_base/src/lib/daemon-client.ts
+++ b/templates/_base/src/lib/daemon-client.ts
@@ -167,7 +167,7 @@ export async function daemonSendFile(
     throw new Error("[volute] daemonSendFile: VOLUTE_DAEMON_PORT or VOLUTE_MIND not set");
   }
   const res = await fetch(
-    `http://127.0.0.1:${port}/api/minds/${encodeURIComponent(mind)}/files/send`,
+    `http://127.0.0.1:${port}/api/v1/minds/${encodeURIComponent(mind)}/files/send`,
     {
       method: "POST",
       headers: headers(),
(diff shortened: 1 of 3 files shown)


hekataion:
Line-by-line review (post-hoc — the PR was opened off a `--stat` check; this is the real pass).

Both call sites correct. Guard test reviewed in full: recursion skips `node_modules`, the violation message carries file:line:path, and the scoping comment correctly explains why widening to `packages/` would false-positive on remote-service calls.

**One known limitation, recorded rather than fixed:** `DAEMON_API_URL` keys on `127.0.0.1:\${...}` and the `/api/...` path being in the *same string literal*. That covers every daemon-bound URL in `templates/` today (all four are full template literals). It would not catch a refactor that splits the base out, e.g.:

```ts
const base = `http://127.0.0.1:\${port}`;
await fetch(`\${base}/api/minds/...`);   // guard blind
```

A stricter rule is available and would have zero false positives right now — *every* `/api/` literal under `templates/` must be `/api/v1/`, since templates only ever address the local daemon. Not making that change on a release-blocker hotfix at 11pm; filing the option here so the choice is deliberate rather than forgotten.

